PR #10, opened 2020-11-18
Descriptors and PSBT used to setup a 2of2 multisig on two wallets

gabridome:
Bitcoin Core and BDK don't know about each other but follow
the same Descriptors and PSBT standards. This allow two users
to set up easily a 2of2 multisig even if they use different
wallets.

NOTE:I'm aware that the structure of the blog is not ready so please feel 
free to review the post until when it has a proper placement.

### DIFF
--- a/layouts/shortcodes/sponsors.html
+++ b/layouts/shortcodes/sponsors.html
@@ -7,7 +7,7 @@
         </a>
     </div>
     <div class="p-3">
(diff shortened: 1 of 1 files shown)


afilini:
Thanks for the contribution!

I'm working on the structure for the blog right now, it shouldn't take too long.. I'll add a comment here with more instructions once I'm done with it!

PR #14, opened 2020-11-19
Add social media preview cards

justinmoon:
I tried tweeting out a link to the new "descriptors in the wild" blog post and it didn't show a nice preview image.

I followed [this post](https://blog.jbowen.dev/2020/06/creating-twitter-cards/) and think I've done everything needed for basic social media cards. But I don't have any way to test it. I tried ngrok.com, but hugo is still using localhost for all the image so it doesn't work. But looks right on inspection.

image

Also, I added a crappy wider jpg version of the logo because the post seemed to suggest that you needed such dimensions and SVG wouldn't work. @afilini I think you might just want to play around with it a bit. Should be pretty straightforward.

[Here's a tool to test Twitter.](https://cards-dev.twitter.com/validator)


afilini:
Awesome, I noticed that too, but I thought it was just my app not loading it or something.. It didn't occur to me that maybe we were doing something wrong :joy:

PR #15, opened 2020-12-03
Automatically publish to GitHub Pages

afilini:
Instead of having to manually build and publish, this workflow automatically builds the website using Hugo whenever something is pushed to `master`, and then commits and pushes the new build to the `gh-pages` branch

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,27 @@
+name: publish
+on:
+  push:
+    branches:
+    - master
+
+jobs:
+  build:
+    name: Build and publish
+    runs-on: ubuntu-latest
+    env:
+      HUGO_VERSION: 0.79.0
+    if: github.ref == 'refs/heads/master'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: '0'
+          submodules: 'recursive'
+      - name: Configure git
+        run: git config user.email "github-actions@github.com" && git config user.name "github-actions"
+      - name: Install Hugo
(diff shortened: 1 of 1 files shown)


RCasatta:
is this also building the wasm code?

afilini:
No, that's still done manually. Mainly because we don't change it that often, so there's no need to rebuilt it every time I guess

.github/workflows/publish.yml
```diff
@@ -0,0 +1,27 @@
...
+          fetch-depth: '0'
+          submodules: 'recursive'
+      - name: Configure git
+        run: git config user.email "github-actions@github.com" && git config user.name "github-actions"
+      - name: Install Hugo
+        run: curl -OL https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/hugo_${HUGO_VERSION}_Linux-64bit.deb && sudo dpkg -i hugo_${HUGO_VERSION}_Linux-64bit.deb && rm -v hugo_*.deb
+      - name: Run build
+        run: ./publish_to_ghpages.sh
+      - name: Push commit
+        run: cd public && git push origin gh-pages
```
notmandatory, line 27: Since we're going to be doing this nightly for the docs I think it makes sense to do a `git --force push origin gh-pages` now.

afilini: I didn't realize this at first, but actually I think leaving the `--force` out is better because it would allow us to catch race condition between different builds. Without `--force` the workflow will fail if the `gh-pages` branch changed while the build was running, maybe because another build pushed something or because we did manually.

At least with a workflow failure we can see that something went wrong, with `--force` it would just silently overwrite the branch

RCasatta:
utack c4130502b5595596fe4ab901f7c9f6fa15c80939, I think we can go ahead and if @notmandatory come up with a counterargument for `--force` we can change it later

.github/workflows/publish.yml
notmandatory: :+1: sounds good leaving force off.

PR #17, opened 2020-12-18
Update the home-page and the rust docs link

afilini:


RCasatta:
utack fd0146177b5354da8303c33c4ef2b62da493b76a,

do we want to add also the link to latest released crates doc?

PR #29, opened 2021-02-23
Add spending policy demo blog post

notmandatory:


content/blog/2021/spending_policy_demo.md
```diff
@@ -0,0 +1,435 @@
...
+export BOB_XPRV=$(cat bob-key.json | jq -r '.xprv')
+
+export CAROL_XPRV=$(cat carol-key.json | jq -r '.xprv')
+```
+
+## Step 3: Derive public extended keys
+
+For this example we are using the [BIP-84](https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki) key path: `m/84h/1h/0h/0/*` to derive extended public keys to share with other wallet participants. 
+
+We also use the `sed` Unix command to remove the origin master key fingerprint and path from our xpub keys since those parts do not need to be shared between wallet participants.
```
afilini: The origin path is not required to generate addresses, but it's good practice to include it because some signers require it. One example are hardware wallets, if one of those keys instead was a `tpub` coming from an hardware wallet, it would most likely require the full path to be able to understand the psbt.

content/blog/2021/spending_policy_demo.md
```diff
@@ -0,0 +1,435 @@
...
+
+We used the [BDK Playground Policy Compiler](https://bitcoindevkit.org/bdk-cli/playground/) to compile the [miniscript](http://bitcoin.sipa.be/miniscript/) policy:
+
+`thresh(3,pk(Alice),pk(Bob),pk(Carol),older(2))`
+
+To the [output descriptor](https://bitcoindevkit.org/descriptors/):
+
+`wsh(thresh(3,pk(Alice),s:pk(Bob),s:pk(Carol),sdv:older(2)))`
+
+This descriptor requires spending transaction inputs must be signed by all three signers, or by two signers ~17mins (512*2 seconds) after being included in a bitcoin block.
```
afilini: Since the `(1 << 22)` bit is not set, this should be in "blocks" mode, not "timestamp" mode. It can be confusing because two blocks take more or less the same time to be mined.

https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki#specification

content/blog/2021/spending_policy_demo.md
```diff
@@ -0,0 +1,435 @@
...
+{
+  "satoshi": 10000
+}
+
+# NO CAROL SHE LOST HER KEY!
+```
+
+## Step 4b: Create spending transaction
+
+This spending transaction uses Alice and Bob's keys plus the ~17 minute (512x2 second) relative time-lock, see above [Step 4a](#step-4a-view-wallet-spending-policies) for the policy id. The transaction can be created by Alice or Bob.
```
afilini: Same as above

afilini (commented):
Can't wait to merge this, I'm sure many people on twitter will love it :)

content/blog/2021/spending_policy_demo.md
notmandatory: Ahh, now I think I get it. So if I want to make this value a relative time lock (512x2 seconds) should I use `sdv:older(4194306)` ? (this is 0x02 with bit 22 also set)

content/blog/2021/spending_policy_demo.md
notmandatory: Oh good point, I'll remove the `sed` part from the text and script parts so we keep the origin paths in the xpubs.

content/blog/2021/spending_policy_demo.md
notmandatory: On second thought I'm going to just reword to explain we're using a two blocks delay as the condition. It's easier to explain, but I'll also include a link to the spec for anyone who wants to play with "timestamp" mode.

afilini: Yes, exactly. Technically not only that specific value because a mask is applied, so even like `(1 << 22) | (1 << 17) | 2` should work, because the mask only covers up to the 15th bit, so `(1 << 17)` would get cancelled.

But yeah, it's probably not a good idea to rely on that behavior since they might use those extra bits for other stuff in the future and break it.

RCasatta:
very nice post!

content/blog/2021/spending_policy_demo.md
thunderbiscuit: I think that for newcomers to wallet descriptors, it's nice to know that the origin path "metadata" part of the descriptor is not necessary (it's sort of obvious why it's nice to have, but it's also not clear if it's mandatory or not at first). Once I got that through trying both ways I felt better/clearer about them. See my comment above for maybe a simple line mentioning that.

content/blog/2021/spending_policy_demo.md
```diff
@@ -0,0 +1,435 @@
...
+export ALICE_XPRV=$(cat alice-key.json | jq -r '.xprv')
+
+export BOB_XPRV=$(cat bob-key.json | jq -r '.xprv')
+
+export CAROL_XPRV=$(cat carol-key.json | jq -r '.xprv')
+```
+
+## Step 3: Derive public extended keys
+
+For this example we are using the [BIP-84](https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki) key path: `m/84h/1h/0h/0/*` to derive extended public keys to share with other wallet participants.
```
thunderbiscuit, line 84: Suggestion for small extra paragraph: 

Note that the `derive` subcommand will generate a `tpub` for the last hardened node in the given derivation path. You'll also notice that `bdk-cli` will returns our `tpub` with the key fingerprint added to it (the metadata part that looks like `[5adb4683/84'/1'/0']` right before the `tpub`). This fingerprint is not necessary in order to use an `tpub` and generate addresses, but it's good practice to include it because some signers require it.

content/blog/2021/spending_policy_demo.md
```diff
@@ -0,0 +1,435 @@
...
+export BOB_DESCRIPTOR="wsh(thresh(3,pk($ALICE_XPUB),s:pk($BOB_XPRV/84'/1'/0'/0/*),s:pk($CAROL_XPUB),sdv:older(2)))" 
+
+export CAROL_DESCRIPTOR="wsh(thresh(3,pk($ALICE_XPUB),s:pk($BOB_XPUB),s:pk($CAROL_XPRV/84'/1'/0'/0/*),sdv:older(2)))"
+```
+
+# Policy A. Three signatures
+
+## Step 1a: Create a testnet [segwit0](https://en.bitcoin.it/wiki/Segregated_Witness) receive address
+
+This step can be done independently by Alice, Bob, or Carol.
```
thunderbiscuit, line 128: Suggesting adding to the sentence:

This step can be done independently by Alice, Bob, or Carol, using any variation of their common descriptor with their `xpub` or `xprv`(for example the same addresses would be derived using a descriptor that contained all 3 `tprv` or one that would contain 3 `tpub`).

_Note: not sure this is really well worded, and might not be necessary (since this post is not about derivations per se so potentially out of scope). I just feel like it's one of the things that trip you up at first. Does the Alice_descriptor, with her `tprv` and their `tpub`, generate the same address as Bob's, who has his `xprv` and their `tpub`? Yes of course, but I had to prove it to myself at first._

notmandatory:
After @thunderbiscuit gives this a test to double check it I'll merge it and make some announcements about the new `bdk-cli` `v0.2.0` release too

thunderbiscuit:
I'm working on it as we speak. Very minor comments.

content/blog/2021/spending_policy_demo.md
```diff
@@ -0,0 +1,435 @@
...
+    "threshold": 3,
+    "type": "THRESH"
+  },
+  "internal": null
+}
+```
+
+## Step 5a: Create spending transaction
+
+The transaction can also be created by Alice, Bob, or Carol.
```
thunderbiscuit: Suggested additon:

The transaction can also be created by Alice, Bob, or Carol, or any third party that has all 3 `tpub`. Note that the argument provided to the `--external_policy` flag contains the id retrieved from the `policies` subcommand, in this case `tjdhrksc`.

content/blog/2021/spending_policy_demo.md
```diff
@@ -0,0 +1,435 @@
...
+    "received": 0,
+    "sent": 10000,
+    "timestamp": 1614058791,
+    "transaction": null,
+    "txid": "3b9a7ac610afc91f1d1a0dd844e609376278fe7210c69b7ef663c5a8e8308f3e"
+  },
+  "psbt": "cHNidP8BAFIBAAAAAYx7T0cL7EoUYBEU0mSL6+DS4VQafUzJgAf0Ftlbkya5AQAAAAD/////AWcmAAAAAAAAFgAU3RacollkleIxk+lz8my/mLCXiH0AAAAAAAEBKxAnAAAAAAAAIgAgCBH16JNfSPhmjJR75EdDB+gSCEF7tStNOLqw/k3BvA0BBXchA3c1Ak2kcGOzOh6eRXFKfpnpzP1lzfcXIYhxFGZG51mxrHwhA75YDXRLDLt+eX5UsE03mIGUSsQP2MrJ9lm17cGXDw2mrJN8IQIvNjaP+mwNC0DtgaB6ENB/DPPlbUDR6+NZ4Sw070jzOKyTfHZjUrJpaJNThyIGAi82No/6bA0LQO2BoHoQ0H8M8+VtQNHr41nhLDTvSPM4DO66tnIAAAAAAAAAACIGA3c1Ak2kcGOzOh6eRXFKfpnpzP1lzfcXIYhxFGZG51mxGFrbRoNUAACAAQAAgAAAAIAAAAAAAAAAACIGA75YDXRLDLt+eX5UsE03mIGUSsQP2MrJ9lm17cGXDw2mDEMxpeYAAAAAAAAAAAAA"
+}
+
+export ALICE_PSBT=$(bdk-cli wallet -w alice -d $ALICE_DESCRIPTOR create_tx -a --to tb1qm5tfegjevj27yvvna9elym9lnzcf0zraxgl8z2:0 --external_policy "{\"tjdhrksc\": [0,1,2]}" | jq -r ".psbt")
```
thunderbiscuit: For me, this variable feels like it should be called something like UNSIGNED_PSBT, because it's the pure, empty psbt. In this step Alice doesn't actually sign it at all, she simply creates it.

content/blog/2021/spending_policy_demo.md
```diff
@@ -0,0 +1,435 @@
...
+```
+
+## Step 6a: Sign and finalize PSBTs
+
+```bash
+# ALICE SIGNS
+export BOB_PSBT=$(bdk-cli wallet -w alice -d $ALICE_DESCRIPTOR sign --psbt $ALICE_PSBT | jq -r ".psbt")
```
thunderbiscuit: Given my comment above, this one could become ALICE_SIGNED_PSBT, which feels more natural as this "object that she then passes on to Bob".

content/blog/2021/spending_policy_demo.md
```diff
@@ -0,0 +1,435 @@
...
+
+# BOB SIGNS
+export CAROL_PSBT=$(bdk-cli wallet -w bob -d $BOB_DESCRIPTOR sign --psbt $BOB_PSBT | jq -r ".psbt")
```
thunderbiscuit: Same as above. This could become BOB_SIGNED_PSBT.

thunderbiscuit (commented):
Great stuff!

content/blog/2021/spending_policy_demo.md
notmandatory: Yes this is good info to add, I just changed it a little to use "key origin" instead of "key fingerprint", since I believe the fingerprint is just the first part.. but when combined with the key path it becomes the xpub key "origin".

content/blog/2021/spending_policy_demo.md
thunderbiscuit: Feel free to disregard this. I'm back and forth on it. It's not of particular importance to the article itself, it's just preemptively answering a question that I would imagine "newish to wallets" readers would have.

notmandatory: This is good info, but I'm going to leave it out of this article to keep it simpler. Since I'm using the same $CAROL_DESCRIPTOR to generate addresses and to do final signing, I don't want to confuse anyone who might think they can somehow sign will all xpubs or something like that.

content/blog/2021/spending_policy_demo.md
notmandatory: This is a good point, and it is common to have an untrusted coordinator who would only have pub keys.

content/blog/2021/spending_policy_demo.md
notmandatory: I see what you mean but my focus was more around naming it for who needs to sign it. I think your point is anyone can sign initially, which is true. Going to just leave it as is for this one.

notmandatory: More descriptive names do make sense.. I'm changing to:
UNSIGNED_PSBT
ALICE_SIGNED_PSBT
ALICE_BOB_SIGNED_PSBT
FINAL_PSBT

PR #40, opened 2021-04-19
updated section1 on seed generation to use bdk-cli instead of ku

i5hi:
* code snippet updates are limited to section 1 on seed generation.

* the output of `bdk-cli derive .. | jq -r .xpub`  is padded with "/* ",  so cut was required to trim it

* a few typos are corrected


i5hi:
Was just looking at the diff and noticed that the `fingerprint` variable is now redundant.

i5hi:
Will update and make a new PR

PR #45, opened 2021-05-21
Fixed typo in miniscript example command

i5hi:
The miniscriptc example code was missing "quotes" and had an additional bracket.


notmandatory:
good catch, verified and merged

PR #53, opened 2021-08-05
Hidden powers of Bitcoin Blog

sandipndev:
In this blog, I have introduced the idea of a descriptor, made a multisig descriptor wallet and tested miniscript policies.
This is a very beginner's take on all these topics.
Hope you guys like it.


notmandatory:
I did a quick read and looks good so far, I'll do a more thorough read tomorrow. In the meantime please add an authors file for yourself in: `content/authors/<your-name>/_index.md`

content/authors/sandipndev/_index.md
```diff
@@ -0,0 +1,8 @@
+---
+name: Sandipan Dey
```
rajarshimaitra: File name should be `sandipan-dey`. Otherwise hugo wont link the author to the page.

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+
+To send people BTC - we simply scan a QR Code *(or paste an address)*, enter some amount and *whoosh* - sent!
+Users might think, just like traditional currency, we can only exchange money using Bitcoin.
+As it so happens, the underlying technology Bitcoin supports specify outputs not as addresses, but as programming scripts.
+This opens us to a world of possibilities using Bitcoin.
+
+### Script
+
+Bitcoin supports [Script](https://en.bitcoin.it/wiki/Script), a **stack-based** lightweight programming language.
+Any script written in **Script** *(pun intended)* contains only `OP_*` codes that Bitcoin Full Nodes understand and process.
```
rajarshimaitra: Script contains both `OP_*` codes and raw byte arrays.

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+Currently, there are `186` op-codes in use.
```
rajarshimaitra: `186` seems too much. As per my counting there are 117 opcodes (including OP_NOPs).  

Also maybe its better to link the wiki page for scripts instead the core code repo? Thats more descriptive and maybe helpful for new learners. https://en.bitcoin.it/wiki/Script

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+You can find a list of the op-codes straight from `bitcoin-core` code repository [here](https://github.com/bitcoin/bitcoin/blob/master/src/script/script.h#L65-L204).
+
+Script is intentionally left [Turing incomplete](https://en.wikipedia.org/wiki/Turing_completeness) which is why there is no [halting problem](https://en.wikipedia.org/wiki/Halting_problem) with scripts.
+There are no loops and overall, it's a very constrained programming language.
+
+A transaction is considered valid only when an Output Script returns `true` at the end of execution.
+Output Scripts define the conditions under which the coins associated with them can be spent.
+Really, to spend the output of a particular coin implies finding an input *(signature)* that makes a particular script evaluate to `true`.
```
rajarshimaitra: Techincally "Output script" doesn't evaluate to true on its own. Below is a re-framing suggestion of the above paragraph. 

 ```suggestion
 A transaction is considered valid only when the Script returns `true` at the end of execution.
Output Script (aka scriptpubkey) define the conditions under which coins associated with them can be spent. To spend a particular coin implies finding an input script (aka scriptsig) such that a script made out of concatenation of `scriptsig + scriptpubkey` evaluates to `true`.
```

File: content/blog/2021/hidden_powers_of_bitcoin.md [hunk 5]
```diff
@@ -0,0 +1,508 @@
...
+scriptPubKey: OP_DUP OP_HASH160 <pubKeyHash> OP_EQUALVERIFY OP_CHECKSIG
+scriptSig: <sig> <pubKey>
+```
+
+##### Examples of things achievable using Bitcoin Script:
+
+1. `Pay Someone (p2pkh/p2wpkh)` - A specific public key must sign to use the coins.
+2. `Escrow (2-of-3-multisig)` - Two parties need to sign together to use the coins.
+3. `Vault (locked)` - A specific key will not be able to use the coins for some time but another master key will always be able to use them.
+4. `HTLC` - The receiver needs to ACK the tx before a time, else the coins are transferred back to the payee.
```
rajarshimaitra: Suggestions inline. 
```suggestion
1. `Pay Someone (p2pkh/p2wpkh)` - A specific public key must sign to spend the coins.
2. `Escrow (2-of-3-multisig)` - Two parties need to sign together to spend the coins.
3. `Vault (locked)` - A specific key will not be able to spend the coins until a timeout but another master key will always be able to spend them.
4. `HTLC` - The receiver needs disclose a secret before a timeout, else the coins are transferred back to the payee.
```

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+
+##### Motivation for Policies
+
+Unfortunately, due to its low-level and unusual stack-based nature, Script is pretty hard to reason about and use.
+Despite being around since Bitcoin creation, writing and understanding Script is not trivial.
```
rajarshimaitra:
```suggestion
Despite being around since Bitcoin's creation, writing and understanding Script is not trivial.
```

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+### Miniscript
+
+[Miniscript](http://bitcoin.sipa.be/miniscript/) tackles the above problems head-on.
+It is an expressive way to create policies on Bitcoin Scripts in a structured and simple fashion.
+Using Miniscript, it's difficult to go wrong.
+
+Another very important goal of Miniscript is to replace any key used in a policy with another policy.
+This is important because people might have multiple keys and complicated timelocks in their existing setup.
+While signing a new policy, they would want to use this existing setup of theirs and also generate addresses for this setup.
+This is something accomplished using something called **Output Descriptors** which we will read about later.
```
rajarshimaitra: This paragraph seemed a little confusing to me. Are you mentioning that we can replace a "key" or "value" inside a policy with another policy? Maybe I am wrong, but is this in general true?

In the last line you mentioned "Descriptors", so were you talking about replacing one "key" with another "key"? 

Nit:
  ```suggestion
Another very important goal of Miniscript is to replace any key used in a policy with another policy.
This is important because people might have multiple keys and complicated timelocks in their existing setup.
While signing a new policy, they would want to use their existing setup to also generate addresses for this new setup.
This is accomplished using something called **Output Descriptors** which we will get into in next section.
```

File: content/blog/2021/hidden_powers_of_bitcoin.md [hunk 8]
```diff
@@ -0,0 +1,508 @@
...
+5. `or(expr_1, expr_2)` - Any one of the expressions should evaluate to true.
+6. `aor(expr_1, expr_2)` - Similar to `or` but `expr_1` has a more probability to evaluate to true.
+
+Bitcoin Script allows us to use another alternate stack. The combinator functions use this second stack to evaluate expressions of `thresh`, `and`, `aor` and `or`.
+The complete Miniscript Reference can be found [here](http://bitcoin.sipa.be/miniscript/#:~:text=Miniscript%20reference).
+
+##### Example Policies
+
+Here are the Miniscript Policies for the examples we looked at earlier. 
+Note `A`, `B`, `C` are placeholders for keys *(or output descriptors)* involved in the tx.
```
rajarshimaitra: You cant put a descriptor inside miniscript, its other way around. 

Descriptors are high level description of scriptpubkey (p2pkh, p2sh etc). And miniscripts are semantics that describes the actual redeemscript. In general you have `Descriptor(Miniscript)` format.

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+Let's understand this with an example of an Output Descriptor:
+
+```
+
+Descriptor: pkh([d34db33f/44'/0'/0']xpub6ERaJH[...]LJRcEL/1/*)#ml40v0wf
+            <1> <--------2---------><----------3---------><4> <---5--->
+
+Sections:
+1 - function type (here, P2PK)
+2 - derivation path of this key
```
rajarshimaitra: Nit:
```suggestion
2 - master key fingerprint and derivation path from master
```

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+3 - xpub at m/44'/0'/0
+4 - path to deriving keys/addresses at
+5 - checksum for the descriptor
+```
+
+This was an example of a simple descriptor using one function but they can also be used in conjugation with multiple functions like `combo`, `multi` etc.
+The concept of descriptor came into existence in 2018 and since then, a lot of wallets have added support for descriptors.
```
rajarshimaitra: Might be a good idea to include a link for descriptor doc here somewhere. https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+Bitcoin uses PSBTs as the standard protocol to create a transaction and add one or more signatures to it before broadcasting the same to 
+the network which finally can become a proper valid *transaction*.
+
+```bash
+# address to send tx to from this multisig wallet
+elcli getnewaddress
+
+# create the transaction, can be started by anyone
+PSBT=$(bdk-cli -n regtest wallet --server tcp://127.0.0.1:60401 -w K2 -d $K2_DESC \
+	create_tx --to "<address-from-last-step>:5555555" | jq -r ".psbt")
```
rajarshimaitra: Instead of sending address from last step its better to store it in a variable. Its done like that everywhere else.

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+bcrt1qptwd6ggy8ttryck2f6yjf4la68apruc3fs7elz
+```
+
+Notes:
+- The `repeat n {}` syntax will only work in `zsh`
+- In case you get different outputs in either of the cases, try deleting `~/.bdk-bitcoin` and retrying (thanks [@Steve](https://twitter.com/notmandatory) for this tip!)
+
+So now we have definitive proof that descriptors can make wallets portable. That single string will be able to make any wallet generate the same set of addresses and hence they can sync and broadcast transactions in the same manner!
+
+#### Making a Secure Descriptor for Funds
```
rajarshimaitra: It seems this a "Multisig" transaction section. "Secure" is generic term and can mean many things.

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+It was pretty insecure because in case any one of those backups leaks, our entire account would be compromised.
+Complicated multisig wallet descriptors are definitely a step forward - just in case a single key leak or are lost, no one would be able to take charge of the funds we hold.
+
+### Retention Bonus - Smart Contract with Bitcoin
+
+Let us consider that a company wants to give its employees a retention bonus for two months.
+If an employee stays with that company for over 2 months, the employee would get 1 BTC as a reward.
+This would be a smart contract between the company and an employee.
+The employee should be able to see that he would get his funds after a year.
+The company would require confidence that the employee would not be able to withdraw the reward before 1 year has passed.
```
rajarshimaitra: Shouldn't the employee get the funds after 2 months (not 1 year)?

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+
+# Delete default wallet created by bitcoindevkit/electrs
+elcli unloadwallet "bdk-test"
+```
+
+Please note that `elcli` is an alias for `bitcoin-cli`.
+
+#### Keys and Generating Addresses
+
+Let us first generate an XPRV and create the wpkh wallet descriptor
```
rajarshimaitra, line 197: Maybe before this step its better to include installation instruction for bdk-cli? 

bdk-cli would also needs a backend to work with. In latest update the default electrum backend is removed. So in order to use bdk-cli for this tutorial they need to compile it with `--features electrum --features compiler`.

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+```
+or(99@and(pk(E),older(8640)),pk(C))
+```
+where `E` is the employee and `C` is the company.
+
+I should emphasize over here that this policy will let the company still transfer funds after the designated 2 months.
+It's not possible to block them after the lock time has passed, atleast not in a single policy.
+
+Surely, after two months, the funds can be unlocked by the employee but before that, the company can revoke the funds.
+Let us compile this policy down to miniscript.
```
rajarshimaitra: The policy is compiled down to a `descriptor`. Not another miniscript.

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+
+```bash
+# The Descriptor will be on the log, the E and C are placeholders
+❯ miniscriptc "or(99@and(pk(E),older(8640)),pk(C))" sh-wsh
```
rajarshimaitra, line 409: This probably old version of bdk-cli. Currently it doesn't produce `miniscriptc` anymore. we have to use `bdk-cli compile` command.

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+
+Let us consider that a company wants to give its employees a retention bonus for two months.
+If an employee stays with that company for over 2 months, the employee would get 1 BTC as a reward.
+This would be a smart contract between the company and an employee.
+The employee should be able to see that he would get his funds after a year.
+The company would require confidence that the employee would not be able to withdraw the reward before 1 year has passed.
+
+The Miniscript policy for this contract would be as follows:
+```
+or(99@and(pk(E),older(8640)),pk(C))
```
rajarshimaitra, line 397: Either I am reading it wrong or 8640 is blocks worth of 1 month, not 2.

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+# The Descriptor will be on the log, the E and C are placeholders
+❯ miniscriptc "or(99@and(pk(E),older(8640)),pk(C))" sh-wsh
+[2021-08-05T12:25:40Z INFO  miniscriptc] Compiling policy: or(99@and(pk(E),older(8640)),pk(C))
+[2021-08-05T12:25:40Z INFO  miniscriptc] ... Descriptor: sh(wsh(andor(pk(E),older(8640),pk(C))))#55wzucxa
+Error: Descriptor(Miniscript(Unexpected("Key too short (<66 char), doesn't match any format")))
+```
+
+So the compiled miniscript is
+```
+sh(wsh(andor(pk(E),older(8640),pk(C))))
```
rajarshimaitra: This is a descriptor. The miniscript is inside the descriptor.

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+In this blog, we will later use the same using [bitcoindevkit/bdk](https://github.com/bitcoindevkit/bdk), a lightweight descriptor-based wallet library
+with a [cli](https://github.com/bitcoindevkit/bdk-cli). 
+
+### Descriptors
+
+Descriptors are a compact and efficient way to "describe" how scripts *(and addresses)* of a wallet should be generated.
+They make it easier to deal with Multisig or complicated key setups.
+Descriptors are super portable and can be easily used by any wallet to determine the list of all addresses that can be generated from the same.
+This feature creates a common stage for all Bitcoin apps and software.
+It's an abstract way to replace keys.
```
rajarshimaitra: To describe a little clearly the difference between a Descriptor and Miniscript, maybe you can add something along this line: 

"A bitcoin scriptpubkey can be of various types, ex P2PKH, P2SH, P2SH-P2WPKH etc. A descriptor is a simple description of what scriptpubkey scheme to be used for a given policy. A Descriptor can include a single pubkey within itself, or an entire miniscript policy. The miniscript policy will be used to derive the resulting redeemscript (the actual executable script), where as the descriptor describes how the redeemscript will be encumbered within the scriptpubkey."

content/blog/2021/hidden_powers_of_bitcoin.md
```diff
@@ -0,0 +1,508 @@
...
+
+Sections:
+1 - function type (here, P2PK)
+2 - derivation path of this key
+3 - xpub at m/44'/0'/0
+4 - path to deriving keys/addresses at
+5 - checksum for the descriptor
+```
+
+This was an example of a simple descriptor using one function but they can also be used in conjugation with multiple functions like `combo`, `multi` etc.
```
rajarshimaitra, line 146: `combo` and `multi` are two different things. `combo` is a descriptor which describes scriptpubkey of any type ex: P2PKH, P2SH etc.

`multi` is basically a miniscript which describes a m-of-n multisig script. When used as descriptor it will make the scriptpubkey as the raw multisig script. But its better to think it as a miniscript than a descriptor, because such raw multisigs are never used in practice.

In wild you will see stuffs like `wsh(multi(2, PKA, PKB, PKC))`. The `wsh` part is the descriptor describing its a P2WSH output, and `multi` part is the miniscript describing the actual spending condition.

Ya it gets messy with terminology in bitcoin scripts. But if we aren't careful, this can cause more confusion to readers.

rajarshimaitra (commented):
Thanks for such a detail walk through of minsicripts with bdk-cli. Its always good to have more miniscript tutorials as its a complicated concept.

untested ACK. 

Below are some comments and suggestions.

content/blog/2021/hidden_powers_of_bitcoin.md
sandipndev: Yes, this was a miscalculation on my part, `117` is correct 😅
And yup, the Wiki Page has better descriptions on each of the OP codes., adding it!

content/blog/2021/hidden_powers_of_bitcoin.md
sandipndev: I tried `cargo install bdk-cli --features electrum --features compiler` but this didn't complete successfully.
It's ending in: 
```
error: no binaries are available for install using the selected features
```

content/blog/2021/hidden_powers_of_bitcoin.md
sandipndev: Block Time = 10 mins
2 Months = 2*30*24*60 mins = 86400 mins
n(Blocks) = 8640

Am I doing it wrong somewhere? 😅

content/blog/2021/hidden_powers_of_bitcoin.md
sandipndev: True, I added this section in the blog because it makes the concept clearer.

content/blog/2021/hidden_powers_of_bitcoin.md
rajarshimaitra: Ah that's because the new version 0.2.1 of bdk-cli hasn't been tagged yet. cc @notmandatory . Currently cargo is trying to install 0.2.0, which requires the `repl` feature and is not included by default.

This works with 0.2.0 `cargo install bdk-cli --features electrum --features compiler --features repl`.

But we should get the latest bdk-cli tagged before releasing this tutorial. Otherwise the instructions can break in future.

content/blog/2021/hidden_powers_of_bitcoin.md
rajarshimaitra: Ah my bad.. I made stupid mistake in my calculation. This is correct.

content/blog/2021/hidden_powers_of_bitcoin.md
rajarshimaitra: Also please confirm this with `v0.2.0` as far as I can see there is no `miniscriptc` binary anymore.

rajarshimaitra:
Can you squash the commits? Will review again after that. Thanks for putting me in co-author, wasn't required.

content/blog/2021/hidden_powers_of_bitcoin.md
rajarshimaitra: Also curious, what version of bdk-cli did you use in the tutorial?

rajarshimaitra (commented):
@sandipndev sorry it took me a long time to get back to this.

I went through the tutorial once again and the text looks good to me. Though I would like some framing change near the descriptor and miniscript explanation section.

The code section doesn't work anymore because `bitcoindevkit:electrs` docker image is removed as we are not maintaining them anymore. This needs to be changed into either local `electrs` or `bitcoind`. I think regtest `bitcoind` would be enough and easy to setup. We should have a separate tutorial for setting up regtest and link it in other tutorials for regtest setup. I think most of the tutorials will use regtest backend.

@ConorOkus This is a really valuable literature and would be a very good addition to our updated doc site. There are not enough descriptor tutorials out there, and this one really showcases the power of bdk for playing with descriptors. We can get this to finish line with little push.

The works needed are following
 - fix some explanation framing.
 - change elctrs docker to bitcoind
 - port to new doc format

I don't wanna bother @sandipndev over this as this has been hanging for quite some time now and the remaining work is mostly logistics.

I am already working on the above fixes and would either do a separate PR with @sandipndev as co-author or just push my changes to this branch, depending on which is easier for porting to new website. @ConorOkus your suggestion here would be helpful.

rajarshimaitra:
I have updated the blog post with latest changes to make the codes work. I seem to not have adequate permission to push to this PR.

So I guess the best thing to do here is open another PR on current master. Will do that soon.

rajarshimaitra:
Updated PR opened at https://github.com/bitcoindevkit/bitcoindevkit.org/pull/82

sandipndev:
Thanks for continuing the work :D

ConorOkus:
Sorry was late to this, great stuff!

PR #68, opened 2021-11-16
Add a blog post about the recent Taproot tx made with BDK

afilini:
<s>Still in draft, I need to add all the relevant links.</s>

------

This is just the first part, since I think the whole thing would have been too long for a single post.

### DIFF
--- a/docs/_blog/first_bdk_taproot_tx.md
+++ b/docs/_blog/first_bdk_taproot_tx.md
@@ -31,14 +31,14 @@ straightforward to understand.
 By Friday night (or rather, early Saturday morning) [I had Taproot key-spend working][first-key-spend], which made me pretty optimistic even though the activation date was actually moving closer, now being forecasted for
 Sunday *morning*.
 
-After a few hours of sleep I went back to work and by early Saturday afternoon [I had Taproot key-spend working as well][first-script-spend]. This left me a few hours to coordinate with some friends and [generate a vanity address][vanity-addr]
+After a few hours of sleep I went back to work and by early Saturday afternoon [I had Taproot script-spend working as well][first-script-spend]. This left me a few hours to coordinate with some friends and [generate a vanity address][vanity-addr]
 to deposit funds into temporarily, as we didn't trust sending them to Taproot addresses before the activation (as they were anyone-can-spend according to the pre-activation rules).
 
 After another pretty short night, I woke up a 5:30 AM on Sunday to monitor the activation. I broadcasted our transactions shortly after 6:00 AM as the activation block was being mined. Unfortunately, the first
 three blocks that were enforcing Taproot rules [didn't include any Taproot transaction][no-taproot-transactions-blocks], which indicates that the miners weren't actually running the new Bitcoin Core 22.0 nodes. The fourth block, mined by `Foundry
 USA` [included my transaction][my-taproot-transaction] and [a few others][few-other-transactions].
 
-In the end our transaction was the third Taproot script-spend in the block, but the first to use the new `OP_CHECKSIGADD` opcode, as the two preceeding it were respectively [a single-sig][taproot-single-sig] and [a 2-of-2 multisig][taproot-bitgo-multisig]
+In the end our transaction was the third Taproot script-spend in the block, but the first to use the new [`OP_CHECKSIGADD`] opcode, as the two preceeding it were respectively [a single-sig][taproot-single-sig] and [a 2-of-2 multisig][taproot-bitgo-multisig]
 script, made with with two `OP_CHECKSIG(VERIFY)`s.
 
 Now, with the context out of the way, we can begin talking about the code!
@@ -48,7 +48,7 @@ Now, with the context out of the way, we can begin talking about the code!
 The first dependency I had to update was [rust-bitcoin]. Most of the taproot stuff were already merged in `master` (altough they hadn't been released yet). One notable missing part was the support for [`BIP371`],
 which is an extension of [`BIP174`], aka the `Partially Signed Bitcoin Transaction` BIP. This new BIP defines a few new fields that are required to properly handle Taproot transactions.
 
-Luckily most of the work had already been done by [sanket1729], so I forked his branch and made only few very minor changes, just to expose a structure that I will have to use later which in his code wasn't `public`.
+Luckily most of the work had already been done by [sanket1729], so I forked his branch and made only few very minor changes, just to expose a structure that I will have to use later which in his code wasn't public.
 
 You can find all the commits mentioned here in [my rust-bitcoin `taproot-testing` branch][rust-bitcoin-taproot-testing].
 
@@ -118,13 +118,13 @@ index 674eeee..3d56cbc 100644
 ```
 
 There isn't much to explain here: I disabled the `missing_docs` lint so that the compiler wouldn't complain about the new public methods that aren't documented.
-Then, I added a getter for the `hash` field of `NodeInfo` and made the strcut itself and a bunch of methods public.
+Then, I added a getter for the `hash` field of `NodeInfo` and made the struct itself and a bunch of methods public.
 
 We will use this structure later to recover the merkle root of a Taproot script tree, given one leaf and the other "hidden" branches.
 
 ## rust-miniscript
 
-Moving on to [rust-miniscript]: once again, most of the work required to support Taproot had already been done by [sanket1729], but this time I was working with very "early" prototype-like code, so I was prepared to
+Moving on to [rust-miniscript]: once again, most of the work required to support Taproot had already been done, but this time I was working with very "early" prototype-like code, so I was prepared to
 make some changes to the code to get it to work how I wanted.
 
 Instead of showing one big diff I will talk about the commits individually, which I think will help making more clear what I was doing.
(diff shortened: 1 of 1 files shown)


docs/_blog/first_bdk_taproot_tx.md
```diff
@@ -0,0 +1,665 @@
...
+I began to wonder how hard it would be to patch BDK and add support for Taproot. I knew most of the work had already been done in our main dependencies, [rust-bitcoin] and [rust-miniscript], and so I decided
+to challenge myself: could I make it in time for the activation?
+
+The following day I started digging into the topic. It didn't help that up until that time I only had a rather "high level" idea of how Taproot worked, but luckily all the BIPs were very well written and
+straightforward to understand.
+
+By Friday night (or rather, early Saturday morning) [I had Taproot key-spend working][first-key-spend], which made me pretty optimistic even though the activation date was actually moving closer, now being forecasted for
+Sunday *morning*.
+
+After a few hours of sleep I went back to work and by early Saturday afternoon [I had Taproot key-spend working as well][first-script-spend]. This left me a few hours to coordinate with some friends and [generate a vanity address][vanity-addr]
```
notmandatory: Did you mean "key-spend" here? or was this when you had the tap "script-spend" working?
```suggestion
After a few hours of sleep I went back to work and by early Saturday afternoon [I had Taproot script-spend working as well][first-script-spend]. This left me a few hours to coordinate with some friends and [generate a vanity address][vanity-addr]
```

docs/_blog/first_bdk_taproot_tx.md
```diff
@@ -0,0 +1,665 @@
...
+to deposit funds into temporarily, as we didn't trust sending them to Taproot addresses before the activation (as they were anyone-can-spend according to the pre-activation rules).
+
+After another pretty short night, I woke up a 5:30 AM on Sunday to monitor the activation. I broadcasted our transactions shortly after 6:00 AM as the activation block was being mined. Unfortunately, the first
+three blocks that were enforcing Taproot rules [didn't include any Taproot transaction][no-taproot-transactions-blocks], which indicates that the miners weren't actually running the new Bitcoin Core 22.0 nodes. The fourth block, mined by `Foundry
+USA` [included my transaction][my-taproot-transaction] and [a few others][few-other-transactions].
+
+In the end our transaction was the third Taproot script-spend in the block, but the first to use the new `OP_CHECKSIGADD` opcode, as the two preceeding it were respectively [a single-sig][taproot-single-sig] and [a 2-of-2 multisig][taproot-bitgo-multisig]
```
notmandatory: Could add a link here to  BIP 342:
```suggestion
In the end our transaction was the third Taproot script-spend in the block, but the first to use the new [`OP_CHECKSIGADD`](https://github.com/bitcoin/bips/blob/master/bip-0342.mediawiki#script-execution) opcode, as the two preceeding it were respectively [a single-sig][taproot-single-sig] and [a 2-of-2 multisig][taproot-bitgo-multisig]
```

docs/_blog/first_bdk_taproot_tx.md
```diff
@@ -0,0 +1,665 @@
...
+-    fn combine(a: Self, b: Self) -> Result<Self, TaprootBuilderError> {
++    pub fn combine(a: Self, b: Self) -> Result<Self, TaprootBuilderError> {
+         let mut all_leaves = Vec::with_capacity(a.leaves.len() + b.leaves.len());
+         for mut a_leaf in a.leaves {
+             a_leaf.merkle_branch.push(b.hash)?; // add hashing partner
+
+```
+
+There isn't much to explain here: I disabled the `missing_docs` lint so that the compiler wouldn't complain about the new public methods that aren't documented.
+Then, I added a getter for the `hash` field of `NodeInfo` and made the strcut itself and a bunch of methods public.
```
notmandatory:
```suggestion
Then, I added a getter for the `hash` field of `NodeInfo` and made the struct itself and a bunch of methods public.
```

docs/_blog/first_bdk_taproot_tx.md
```diff
@@ -0,0 +1,665 @@
...
+diff --git a/src/descriptor/tr.rs b/src/descriptor/tr.rs
+index 8487d56..fabf860 100644
+--- a/src/descriptor/tr.rs
++++ b/src/descriptor/tr.rs
+@@ -593,7 +593,7 @@ impl<Pk: MiniscriptKey> DescriptorTrait<Pk> for Tr<Pk> {
+     }
+
+     fn max_satisfaction_weight(&self) -> Result<usize, Error> {
+-        let mut max_wieght = None;
++        let mut max_wieght = Some(65);
```
notmandatory, line 446: Note for your final PR.. the correct spelling is `weight`... the rule "i before e except after c" doesn't apply.. the English language hates it's users. Looks like you have it right everywhere else but code you're fixing was wrong.
```suggestion
+        let mut max_weight = Some(65);
```

docs/_blog/first_bdk_taproot_tx.md
```diff
@@ -0,0 +1,665 @@
...
+```
+
+And finally, the last little fix: the `multi_a()` operator is satisfied by pushing to the witness either a signature (if you have one available for that specific public key) or an empty stack element. The problem is,
+they have to be in the right order to match the order of public keys in your Taproot script.
+
+rust-miniscript was pushing them in reverse order, so script validation was always failing for multisigs that had more than 1 key. Adding a `.rev()` to the iterator fixed the issue.
+
+## Conclusion
+
+And that was it! We now have a fully working [rust-bitcoin] and rust-miniscript ready for Taproot.
```
notmandatory:
```suggestion
And that was it! We now have a fully working [rust-bitcoin] and [rust-miniscript] ready for Taproot.
```

notmandatory:
Great write up! what a journey! I made some small corrections/suggestions.

docs/_blog/first_bdk_taproot_tx.md
afilini: Yeah it was in the original code I was working on. I think I'll leave it like that in this article so that it matches the real code perfectly, but I will point it out during review on rust-miniscript

docs/_blog/first_bdk_taproot_tx.md
```diff
@@ -0,0 +1,665 @@
...
+```rust
+/// Access the internal script map
+pub fn as_script_map(&self) -> &BTreeMap<(Script, LeafVersion), BTreeSet<TaprootMerkleBranch>> {}
+```
+
+Then the code would look for the "shortest" path to that specific script, as it would save size in the final transaction (leaves that are more "deep" in the tree than others naturally have more hidden branches
+in their path to the root, and thus require a longer control block to reveal them all).
+
+The issue here is that the `control_block` variable is then serialized directly into the witness. But this is not a control block, it's just a set of merkle paths! A control block only has *one* merkle path, and
+incldues the leaf version and the key parity bit.
```
RCasatta: *includes

docs/_blog/first_bdk_taproot_tx.md
```diff
@@ -0,0 +1,665 @@
...
+to deposit funds into temporarily, as we didn't trust sending them to Taproot addresses before the activation (as they were anyone-can-spend according to the pre-activation rules).
+
+After another pretty short night, I woke up a 5:30 AM on Sunday to monitor the activation. I broadcasted our transactions shortly after 6:00 AM as the activation block was being mined. Unfortunately, the first
+three blocks that were enforcing Taproot rules [didn't include any Taproot transaction][no-taproot-transactions-blocks], which indicates that the miners weren't actually running the new Bitcoin Core 22.0 nodes. The fourth block, mined by `Foundry
+USA` [included my transaction][my-taproot-transaction] and [a few others][few-other-transactions].
+
+In the end our transaction was the third Taproot script-spend in the block, but the first to use the new [`OP_CHECKSIGADD`] opcode, as the two preceeding it were respectively [a single-sig][taproot-single-sig] and [a 2-of-2 multisig][taproot-bitgo-multisig]
```
RCasatta: *preceding

docs/_blog/first_bdk_taproot_tx.md
```diff
@@ -0,0 +1,665 @@
...
+-                        .iter()
+-                        .min_by(|x, y| x.as_inner().len().cmp(&y.as_inner().len()))
+-                        .expect("Atleast one control must exist for a known leaf");
+                     wit.push(control_block.serialize());
+                     // Finally, save the minimum
+                     min_wit = Some(wit);
+
+```
+
+This is where things get more interesting: this section of code builds the witness to satisfy a Taproot descriptor. In case of a script-spend, we need to prove that the script we are using had been commited
```
RCasatta: *committed

docs/_blog/first_bdk_taproot_tx.md
```diff
@@ -0,0 +1,665 @@
...
+             let script_size = ms.script_size();
+             let max_sat_elems = match ms.max_satisfaction_witness_elements() {
+```
+
+This is a little bug in the code that tries to compute what the maximum satisfaction weight would be for a descriptor. For instance, we use this in BDK to compute how many extra sats of fees we need to pay
+in order to target a given fee rate, assuming the descriptor is satisfied with the worst (larger and most expensive) path.
+
+For Taproot descriptors, it's just a matter of iterating over the leaves in the tree and pick the most expensive one... or is it? This doesn't take into account that Taproot outputs can also be spent with
+key-spend, which means just pushing a signature to the witness. This signature is 64 bytes long when using the new [`SIGHASH_DEFAULT`][`BIP341`] sighash, or 65 otherwise. Since we are thinking about the maximum satisfaction
+weight, or the worst case possible, we natuarlly pick the latter.
```
RCasatta: *naturally

afilini:
Thanks for the review @RCasatta, next time I'll run it through a typo-checker thing before opening the pr..

notmandatory (approved):
LGTM, ready to merge?

afilini:
Yeah, I think so!

PR #74, opened 2021-12-18
descriptors: correct 'pkh()' in the compatibility matrix

darosior:
Just a nit as i read through the compatibility matrix. See the commit message: the table doesn't list `pk_k`, so just keep it to the `pk()` and `pkh()` aliases which are both compatible with legacy descriptors and valid top-level Miniscripts (base expression).


notmandatory (approved):
ACK , makes sense to me.

PR #77, opened 2021-12-24
Add descriptor code sample

w0xlt:
There's a `Descriptor` section in `Getting Started` page, but there is no code showing how to use descriptors with BDK.

This PR adds a code example. This is based on this [video](https://www.youtube.com/watch?v=GHSIPltK-mQ).

But the video uses BDK 0.10.
This code uses the latest version, 0.15.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -29,3 +29,53 @@ The author of this project strongly believes descriptors will be a big part of t
 technology and tooling of Bitcoin evolves and changes (Schnorr signatures, Taproot, etc).
 
 To learn more, check out the specific [Descriptors section](/descriptors).
+
+The following code shows how to generate a random mnemonic, the extended (and deterministic) keys from that mnemonic and finally the descriptors from the extended private keys.
+
+To be able to run this code, the `bdk` dependency in `Cargo.toml` must be set as follows:
+
+```
+[dependencies]
+bdk = { version = "0.15.0", default-feature = false, features = ["all-keys"] }
+```
+
+```rust
+use bdk::bitcoin::Network;
+use bdk::database::MemoryDatabase;
+use bdk::keys::{DerivableKey, GeneratableKey, GeneratedKey, ExtendedKey, bip39::{Mnemonic, WordCount, Language}};
+use bdk::template::Bip84;
+use bdk::{miniscript, Wallet, KeychainKind};
+
+fn main() {
+    println!("Hello, world!");
+
+    let network = Network::Testnet; // Or this can be Network::Bitcoin, Network::Signet or Network::Regtest
+
+    // Generate fresh mnemonic
+    let mnemonic: GeneratedKey<_, miniscript::Segwitv0> = Mnemonic::generate((WordCount::Words12, Language::English)).unwrap();
+    // Convert mnemonic to string
+    let mnemonic_words = mnemonic.to_string();
+    // Parse a mnemonic
+    let mnemonic  = Mnemonic::parse(&mnemonic_words).unwrap();
+    // Generate the extended key
+    let xkey: ExtendedKey = mnemonic.into_extended_key().unwrap();
+    // Get xprv from the extended key
+    let xprv = xkey.into_xprv(network).unwrap();
+
+    // Create a BDK wallet structure using BIP 84 descriptor ("m/84h/1h/0h/0" and "m/84h/1h/0h/1")
+    let wallet = Wallet::new_offline(
+        Bip84(xprv, KeychainKind::External),
+        Some(Bip84(xprv, KeychainKind::Internal)),
+        network,
+        MemoryDatabase::default(),
+    )
+    .unwrap();
+
+    println!("mnemonic: {}\n\nrecv desc (pub key): {:#?}\n\nchng desc (pub key): {:#?}",
+        mnemonic_words,
+        wallet.get_descriptor_for_keychain(KeychainKind::External).to_string(),
+        wallet.get_descriptor_for_keychain(KeychainKind::Internal).to_string());
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -29,3 +29,79 @@ The author of this project strongly believes descriptors will be a big part of t
...
+    for path in ["m/84h/1h/0h/0", "m/84h/1h/0h/1"] {
+        let derived_path = DerivationPath::from_str(path).unwrap();
+        let derived_xpriv = &xprv.derive_priv(&secp, &derived_path).unwrap();
+        let origin: KeySource = (xprv.fingerprint(&secp), derived_path);
+        let derived_xprv_descriptor_key: DescriptorKey<Segwitv0> =
+            derived_xpriv.into_descriptor_key(Some(origin), DerivationPath::default()).unwrap();
+
+
+        // Wrap the derived key with the "wpkh()" string to produce a descriptor text
+        if let Secret(key, _, _) = derived_xprv_descriptor_key {
```
notmandatory: A bit cleaner way to do this is to use the bip84 template, details here: 
https://docs.rs/bdk/0.15.0/bdk/descriptor/template/struct.Bip84.html

docs/getting-started.md
```diff
@@ -29,3 +29,79 @@ The author of this project strongly believes descriptors will be a big part of t
...
 
 To learn more, check out the specific [Descriptors section](/descriptors).
+
+The following code shows how to generate a random mnemonic, the extended (and deterministic) keys from that mnemonic and finally the descriptors from the extended private keys.
+
+To be able to run this code, the `bdk` dependency in `Cargo.toml` must be set as follows:
+
+```
+[dependencies]
+bdk = { version = "0.15.0", default-feature = false, features = ["compiler", "all-keys"] }
```
rajarshimaitra: I think you don't need `compiler` here for just the descriptor generation code.

rajarshimaitra (changes requested):
Concept ACK https://github.com/bitcoindevkit/bitcoindevkit.org/pull/77/commits/4d3c467b9df7676e45d164d232cbda5bba6b3f26

It does make sense to expose the descriptor generation code in getting started page. 

I found few more improvements that can be made in other part of the doc. Probably will open a separate PR for that.

Tested that the code works.

Just one nit.

docs/getting-started.md
rajarshimaitra: I think the example in Bip84 template is useful to show how we can use `bitcoin::utill` to do the same thing. Where as here we wanna show how to use bdk api to do it. I was looking for the same and ended up doing it this way in my rpc tutorial, to only use things exposed in bdk.

I am still not fully satisfied with the amount of code it needs to create descriptors. And probably there's room to compact this out further.

docs/getting-started.md
w0xlt: Thanks. Removed.

docs/getting-started.md
w0xlt: I also have another version with the code below:

```rust
pub fn mnemonic_to_xprv(network: &Network, mnemonic_words: &str) -> ExtendedPrivKey {
    let mnemonic  = Mnemonic::parse(mnemonic_words).unwrap();
    let xkey: ExtendedKey = mnemonic.into_extended_key().unwrap();
    let xprv = xkey.into_xprv(*network).unwrap();
    xprv
}

let wallet_name = wallet_name_from_descriptor(
    Bip84(xpriv.clone(), KeychainKind::External),
    Some(Bip84(*xpriv, KeychainKind::Internal)),
    *network,
    &Secp256k1::new()
 ).unwrap();
    
println!("mnemonic: {}\n\nrecv desc (pub key): {:#?}\n\nchng desc (pub key): {:#?}",
    mnemonic_words,
    wallet.get_descriptor_for_keychain(KeychainKind::External).to_string(),
    wallet.get_descriptor_for_keychain(KeychainKind::Internal).to_string());
 ```
 
I can change this PR to this version, but this one requires a wallet instance to be created before the descriptors are extracted.

rajarshimaitra: Ah I see. I think I missunderstood @notmandatory 's comment. Yes it makes sense to use `Bip84` template derivation instead of manually wrapping in `wpkh` string. 

This will be a good modification to have..

rajarshimaitra (commented):
tACK https://github.com/bitcoindevkit/bitcoindevkit.org/pull/77/commits/e2da710c3fd357093198d76a22d58af0455c68e2

Changes looks good. Would be great to have this code in the tutorial https://github.com/bitcoindevkit/bitcoindevkit.org/pull/77#discussion_r777157086.

docs/getting-started.md
w0xlt: I updated the PR with this code in 60b72fe

w0xlt:
PR updated with the code suggested in https://github.com/bitcoindevkit/bitcoindevkit.org/pull/77#discussion_r776495931

rajarshimaitra (approved):
ACK https://github.com/bitcoindevkit/bitcoindevkit.org/pull/77/commits/60b72fedb9660917d62f08a769cc919f147af403

ConorOkus:
Is it possible to add this to the `examples` directory and make it part of CI?

w0xlt:
@ConorOkus I added this code to the `examples` directory in https://github.com/bitcoindevkit/bdk/pull/526

notmandatory:
Looks like this one is ready to go but needs a final rebase. Sorry for the long delay!

w0xlt:
No problem. Rebased.

notmandatory (approved):
utACK 734a74cd1f6cb0a513b109bbe4d0dd4b363f9917

PR #75, opened 2021-12-23
Update installation.md for different esplora features

BitcoinABee:
We must now specify if you want the blocking esplora client based on ureq or the async version based on reqwest.

This is a fix for issue https://github.com/bitcoindevkit/bitcoindevkit.org/issues/64 and https://github.com/bitcoindevkit/bdk-cli/issues/53

PR #92, opened 2022-03-17
Fix bdk-cli default features install

### LINKED ISSUE #30: Remove the unnecessary --features electrum,repl in the cli instructions
They are now the default as of bitcoindevkit/bdk-cli#22

thunderbiscuit:
Fixes #30.

### DIFF
--- a/docs/_blog/spending_policy_demo.md
+++ b/docs/_blog/spending_policy_demo.md
@@ -3,7 +3,7 @@ title: "Spending Policy Demo"
 description: "Demonstrate how to use a descriptor wallet with different spending policies"
 authors:
     - Steve Myers
-    - Thunderbiscuit
+    - thunderbiscuit
 date: "2021-02-23"
 tags: ["guide", "descriptor"]
 permalink: "/blog/2021/02/spending-policy-demo/"
@@ -31,11 +31,11 @@ In a real-world wallet a longer relative time-lock would probably be used, but w
 ### Step 0: Install a recent version `bdk-cli`
 
 ```bash
-cargo install bdk-cli --features repl,electrum,esplora
+cargo install bdk-cli --features electrum
 
 # confirm bdk-cli is installed
 bdk-cli --version
-BDK CLI 0.2.0
+BDK CLI 0.4.0
 
 # bdk-cli usage can be explored with the `help` sub-command
 bdk-cli help
--- a/docs/bdk-cli/installation.md
+++ b/docs/bdk-cli/installation.md
@@ -32,7 +32,7 @@ cargo install --git https://github.com/bitcoindevkit/bdk-cli --features=esplora-
 # all features with the async esplora client 
 cargo install --git https://github.com/bitcoindevkit/bdk-cli --features=esplora-reqwest,compiler
 
-# minimal install
+# minimal install (only repl feature is on by default)
 cargo install --git https://github.com/bitcoindevkit/bdk-cli
 ```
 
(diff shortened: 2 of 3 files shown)


docs/_blog/spending_policy_demo.md
```diff
@@ -31,11 +31,11 @@ In a real-world wallet a longer relative time-lock would probably be used, but w
 ### Step 0: Install a recent version `bdk-cli`
 
 ```bash
-cargo install bdk-cli --features repl,electrum,esplora
+cargo install bdk-cli --features esplora
```
rajarshimaitra: I did not get why `esplora` is required here.. Electrum is enough for the example it seems??

rajarshimaitra (commented):
ACK 7773339e246c23d14bfdeeede62ee32bcd18e558

With one nit.

docs/_blog/spending_policy_demo.md
notmandatory: agreed should be fine to use `--features electrum` .. but if you do want to use esplora then it needs to be `--features esplora-ureq` or `--features esplora-reqwest`

thunderbiscuit: Ok so then I think I should remove the features completely right? Alekos' comment was that now the electrum feature is a default so we should remove it.

thunderbiscuit:
@rajarshimaitra @notmandatory now ready to go. The line doesn't have any features at all, which the blog post does not require.

```sh
cargo install bdk-cli
```

_Edit:_ This comment is no longer true, the line is `cargo install bdk-cli --features electrum`.

docs/_blog/spending_policy_demo.md
```diff
@@ -31,11 +31,11 @@ In a real-world wallet a longer relative time-lock would probably be used, but w
 ### Step 0: Install a recent version `bdk-cli`
 
 ```bash
-cargo install bdk-cli --features repl,electrum,esplora
+cargo install bdk-cli
```
notmandatory: We do need at least one blockchain feature added, otherwise we can't use any online commands like `sync`.
```suggestion
cargo install bdk-cli --features electrum
```

thunderbiscuit: Oh sorry I thought that the `electrum` feature was added by default?

notmandatory: Ah ya the original issue is based on older version of bdk-cli where the `electrum` feature was enabled by default. But when we added the ability to have different blockchains like esplora or rpc we had to remove it as a default. I hope in the future we'll be able to have all features enabled without conflicting, then we can make electrum added by default again. See https://github.com/bitcoindevkit/bdk-cli/issues/75.

thunderbiscuit: I realize the default electrum feature was actually removed by #75. Re-added.

docs/bdk-cli/installation.md
```diff
@@ -32,7 +32,7 @@ cargo install --git https://github.com/bitcoindevkit/bdk-cli --features=esplora-
 # all features with the async esplora client 
 cargo install --git https://github.com/bitcoindevkit/bdk-cli --features=esplora-reqwest,compiler
 
-# minimal install
+# minimal install (electrum and repl features are default)
```
notmandatory: Sorry needs to be fixed here too:
```suggestion
# minimal install (only repl  feature is on by default)
```

thunderbiscuit: Good catch. Done!

rajarshimaitra (approved):
ReACK b889ded078b4cf5060e7e1c37f1b3fc095e572c7

notmandatory (approved):
ACK b889ded078b4cf5060e7e1c37f1b3fc095e572c7

PR #93, opened 2022-03-20
Tutorial tweaks

mplsgrant:
I updated the tutorial with changes appropriate to the  0.17 release.

Incidentally, I think it may be worth it to ask the user to populate their wallet with testnet coins, otherwise they will get a dust error.

### DIFF
--- a/docs/tutorials/hello-world.md
+++ b/docs/tutorials/hello-world.md
@@ -2,8 +2,7 @@
 
 ## Introduction
 
-This article should serve as a "getting started" guide for developers who are considering integrating BDK in their projects: it tries to introduce the reader to the basic concepts behind the library and some of its
-modules and components that can be used to build a very simple functioning Bitcoin wallet. All the information written in this article are valid for the latest [published version](https://docs.rs/bdk/latest/bdk/).
+This article should serve as a "getting started" guide for developers who are considering integrating BDK in their projects: it tries to introduce the reader to the basic concepts behind the library and some of its modules and components that can be used to build a very simple functioning Bitcoin wallet. All the information written in this article are valid for the latest [published version](https://docs.rs/bdk/latest/bdk/).
 
 ## Design Goals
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/hello-world.md
```diff
@@ -81,7 +81,7 @@ it doesn't know which UTXOs belong to the wallet. To get this information we gen
 To fix this we can add one more component to our `Wallet`: a [`Blockchain`][Blockchain] backend. In particular, we are going to use the [`ElectrumBlockchain`][ElectrumBlockchain] which syncs with an `Electrum` server,
```
rajarshimaitra: This line needs to be changed into something like this.

```suggestion
To fix this we can add one more component in our code: a [`Blockchain`][Blockchain] backend. In particular, we are going to use the [`ElectrumBlockchain`][ElectrumBlockchain] which syncs with an `Electrum` server. And use this blockchain to `sync` our wallet database with current state of the network.
```

docs/tutorials/hello-world.md
```diff
@@ -94,16 +94,11 @@ use bdk::{
...
-let wallet: Wallet<ElectrumBlockchain, MemoryDatabase> = Wallet::new(
-    external_descriptor,
-    Some(internal_descriptor),
-    Network::Testnet,
-    MemoryDatabase::new(),
-    ElectrumBlockchain::from(Client::new("ssl://electrum.blockstream.info:60002").unwrap()),
-)?;
+    let client = Client::new("ssl://electrum.blockstream.info:60002")?;
+    let blockchain = ElectrumBlockchain::from(client);
+
```
rajarshimaitra: Remove extra new line..

docs/tutorials/hello-world.md
```diff
@@ -94,16 +94,11 @@ use bdk::{
...
-)?;
+    let client = Client::new("ssl://electrum.blockstream.info:60002")?;
+    let blockchain = ElectrumBlockchain::from(client);
+        
 ```
 
-This piece of code is very similar to the one we wrote before, but this time we are using the [`Wallet::new()`][Wallet_new] constructor instead of [`Wallet::new_offline()`][Wallet_new_offline], and this takes an extra argument for the `Blockchain` type to use.
 Specifically here, we create an `ElectrumBlockchain` and connect to Blockstream's public Electrum Testnet servers over SSL.
 
 Now, since we are running in the `Testnet` network, we can try to get some funds from a faucet online to this address we've generated. Once we have an incoming transaction we can do the first `sync` of our *online* wallet.
```
rajarshimaitra:
```suggestion
Now, since we are running in the `Testnet` network, we can try to get some funds from a faucet online to this address we've generated. Once we have an incoming transaction we can do the first `sync` of our wallet.
```

docs/tutorials/hello-world.md
```diff
@@ -113,22 +108,19 @@ battery. Having a thread running and trying to make network requests while the a
...
 
 // ...
 
-wallet.sync(noop_progress(), None)?;
+wallet.sync(&blockchain, SyncOptions::default())?;
 ```
 
-In this case, we are not interested in receiving updates about the progress, and we just want to use the default settings, so we use [`noop_progress()`][noop_progress] and `None` as arguments. This will make queries to the Electrum server
-and store the list of transactions and UTXOs in our `Database`. In this case, we are using a `MemoryDatabase`, so those data are only going to be kept in RAM and dropped once our `Wallet` is dropped. This is very useful
+This will make queries to the Electrum server and store the list of transactions and UTXOs in our `Database`. In this case, we are using a `MemoryDatabase`, so those data are only going to be kept in RAM and dropped once our `Wallet` is dropped. This is very useful
```
rajarshimaitra:
```suggestion
The `SyncOptions` determines some sync time behaviors, like progress update, etc. For this case the `default` sync option with no progress update is adequate. This will make queries to the Electrum server and store the list of transactions and UTXOs in our `Database`. In this case, we are using a `MemoryDatabase`, so those data are only going to be kept in RAM and dropped once our `Wallet` is dropped. This is very useful
```

docs/tutorials/hello-world.md
```diff
@@ -180,9 +175,7 @@ println!(
 
 ## Custom Database and Blockchain types
 
-We briefly mentioned before that for our example we used the `MemoryDatabase`, but that it could also be swapped for a different one: this is one example of the *modularity* of BDK. By default, some database
-types are implemented in the library, namely (as of now) the `MemoryDatabase` which only keeps data in RAM and the [sled][sled] database that can store data on a filesystem. But since the `Database` trait is public,
-users of the library can also implement different database types more suitable for their use-case.
+We briefly mentioned before that for our example we used the `MemoryDatabase`, but that it could also be swapped for a different one: this is one example of the *modularity* of BDK. By default, some database types are implemented in the library, namely (as of now) the `MemoryDatabase` which only keeps data in RAM and the [sled][sled] database that can store data on a filesystem. But since the `Database` trait is public, users of the library can also implement different database types more suitable for their use-case.
```
rajarshimaitra: Add `[rusqlite]: https://docs.rs/rusqlite/latest/rusqlite/` after line 193.


```suggestion
We briefly mentioned before that for our example we used the `MemoryDatabase`, but that it could also be swapped for a different one: this is one example of the *modularity* of BDK. By default, some database types are already implemented in the library, namely the `MemoryDatabase` which only keeps data in RAM, the [sled][sled] database that can store data on a filesystem, and the `SqliteDatabase`[rusqlite] that can store data into a SQLite database. But since the `Database` trait is public, users of the library can also implement different database types more suitable for their use-case.
```

docs/tutorials/hello-world.md
```diff
@@ -180,9 +175,7 @@ println!(
...
 
 The same is true for the `Blockchain` types: the library provides (through the use of opt-in features) implementations for the `Electrum`, `Esplora` and `CompactFilters` (*Neutrino*) backends. Those again can also be
```
rajarshimaitra:
```suggestion
The same is true for the `Blockchain` types: the library provides (through the use of opt-in features) implementations for the `Electrum`, `Esplora`, `CompactFilters` (*Neutrino*) and Bitcoin Core `rpc` backends. Those again can also be
```

rajarshimaitra (changes requested):
Initial Concept ACK. Thanks for working on this. Below are some few modifications needed to be made. 

There is a line in the doc like this (line 19)
> Throughout the article, I will try to focus on those points and try to explain them as best as I can.

Its better to remove 1st person singular statements as the project is getting more matured. You can try changing the "I" into "we" here and elsewhere.. For this one specifically I think its better to just remove it. It made sense when Alekos wrote it in the initial days.. 

> This is how an `OfflineWallet` can be created.

Line 38. We don't create an `OfflineWallet` anymore. So this line can be removed. 

> this is again something that might seem counterintuitive at first

Line 105. Make `This` capitalize.

mplsgrant:
I like your suggestions. Thank you! I don't know if I'll have time to take another pass at it this week. *Edit* Though I was able to make a few tweaks.

Is there something you would like from me in the immediacy, perhaps squashing the commits?

rajarshimaitra:
Thanks @mplsgrant I will take a look at it again.. Yes a squash would be helpful here..

docs/tutorials/hello-world.md
```diff
@@ -16,24 +15,18 @@ The main goal of the library is to be a solid foundation for Bitcoin wallets of
...
 
-Going back to our `Wallet`: given a descriptor and a `Database` we can build an "air-gapped", or "Offline" wallet: basically, a wallet that physically can't to connect to the Bitcoin network. It will still be able to generate addresses and
-sign [PSBTs][PSBT], but with a greatly reduced attack surface because a sizable part of the code that handles the logic to synchronize with the network would be entirely omitted in the final executable binary.
+Going back to our `Wallet`: given a descriptor and a `Database` we can build an "air-gapped" or "Offline" wallet. Basically, we can make a wallet that physically can't to connect to the Bitcoin network. It will still be able to generate addresses and sign [PSBTs][PSBT], but with a greatly reduced attack surface because a sizable part of the code that handles the logic to synchronize with the network would be entirely omitted in the final executable binary.
```
rajarshimaitra:
```suggestion
Going back to our `Wallet`: given a descriptor and a `Database` we can build an "air-gapped" or "Offline" wallet. Basically, we can make a wallet that physically can't connect to the Bitcoin network. It will still be able to generate addresses and sign [PSBTs][PSBT], but with a greatly reduced attack surface because a sizable part of the code that handles the logic to synchronize with the network would be entirely omitted in the final executable binary.
```

docs/tutorials/hello-world.md
```diff
@@ -180,17 +160,14 @@ println!(
 
 ## Custom Database and Blockchain types
 
-We briefly mentioned before that for our example we used the `MemoryDatabase`, but that it could also be swapped for a different one: this is one example of the *modularity* of BDK. By default, some database
-types are implemented in the library, namely (as of now) the `MemoryDatabase` which only keeps data in RAM and the [sled][sled] database that can store data on a filesystem. But since the `Database` trait is public,
-users of the library can also implement different database types more suitable for their use-case.
+We briefly mentioned before that for our example we used the `MemoryDatabase`, but that it could also be swapped for a different one: this is one example of the *modularity* of BDK. By default, some database types are already implemented in the library, namely the `MemoryDatabase` which only keeps data in RAM, the [sled][sled] database that can store data on a filesystem, and the `SqliteDatabase`[rusqlite] that can store data into a SQLite database. But since the `Database` trait is public, users of the library can also implement different database types more suitable for their use-case.
```
rajarshimaitra: The `rusqlite` link is missing..

rajarshimaitra (changes requested):
Looks good.. Few more nits.. 

> Line 38. We don't create an OfflineWallet anymore. So this line can be removed.

This one is missed..

docs/tutorials/hello-world.md
mplsgrant: The rusqlite link misbehaved for me. Don't have time to fix the link. So, I removed the link.

mplsgrant:
Committed the changes and squashed.

After removing the mention of OfflineWallet, I noticed the links at the bottom of the page are probably all outdated and also mention OfflineWallet. Will need to fix that at some point.

I'm getting the feeling that it would be better to get this version of the tutorial out in the wild before too many devs get discouraged by a broken hello-world tutorial.

rajarshimaitra:
> After removing the mention of OfflineWallet, I noticed the links at the bottom of the page are probably all outdated and also mention OfflineWallet. Will need to fix that at some point.

Thanks for catching.. Yes these are broken now.. 

> I'm getting the feeling that it would be better to get this version of the tutorial out in the wild before too many devs get discouraged by a broken hello-world tutorial.

Yes I also think this needs to be done on priority basis.. Is it ok if I redo on top of your work with co-author attribution to you into another PR? I can get this fixed faster..

mplsgrant:
>Is it ok if I redo on top of your work with co-author attribution to you into another PR? I can get this fixed faster..

Yes, that sounds great. I appreciate the attribution, and thank you for seeing this through!

rajarshimaitra:
This can now be closed for #94

PR #110, opened 2022-07-27
[bdk_core] Fix broken links

LLFourn:
Use permalinks otherwise the links will break.


notmandatory (approved):
ACK fd23b618a674b7ee200755df60fc26bfb0d348cb

PR #112, opened 2022-08-10
Update playground CSS

danielabrozzoni:
- Kinda-align things together
- Set a fixed height for the iframe, so that when playing with the
  policies there's no scrollbar. It would be awesome if we could
  adjust the iframe height to the content height (so that when new
  things appear, there's no need for a scrollbar), but I don't know
  how I would do that
- Include fontawesome for the + icon

### DIFF
--- a/docs/.vuepress/public/bdk-cli/playground/playground.html
+++ b/docs/.vuepress/public/bdk-cli/playground/playground.html
@@ -4,6 +4,7 @@
   }
 </style>
 
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.1.2/css/all.min.css" integrity="sha512-1sCRPdkRXhBV2PBLUdRb4tMg1w2YPf37qatUFeS7zlBy7jJI8Lf4VHwWfZZfpXtYSLy85pkm9GaYVYMfw5BC1A==" crossorigin="anonymous" referrerpolicy="no-referrer" />
 <div class="expand">
   <div class="expand-label" style="cursor: pointer;"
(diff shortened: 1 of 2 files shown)


rajarshimaitra (approved):
Big ACK 94deab714d4d33477444c6a02591b65f5baab678

Thanks for the CSS love.. This really does make a lot of difference and the playground is looking really nice.. 

The code changes looks good to me and this is ready to go IMO.. 

Few UI Wish (Can be done in later PRs): 

 - A Copy button beside the final compiled descriptor would be really useful.
 - If the output descriptor bar can be multi line with wrapping, I feel thats more easier to read at glance bigger descriptors than as single line with horizontal scroll..
 -  If the Wallet output background can be turned into something different than matching with main page background color, I feel that could be easier to read, and separate it out like a different section..
 - Theres seems to be a different vertical scroll going on in the Playground window, separate from the main page.. Can these be combined into one?? Or like don't have the playground scroll separately from the main window?
 - A little Bit of description at the top of the age how to use this playground, and explaining they can try out a full BDK wallet with their custom policies, could be very helpful..   
 
 
 Also, should we post this is as CSS help wanted issues?? I have no clue how to do these CSS magic myself..

rajarshimaitra:
I was wandering if you have tried updating the playground dep to latest bdk-cli with wasm bindings?? Or else I will give it a shot..

With the latest bdk-cli playing the wallet with a background regtest node in the web could be really cool..

danielabrozzoni:
Yep, I'm still working on that but I'm really near the solution! I hope I'll be able to post a PR soon :)

danielabrozzoni:
Also, rebased

notmandatory (approved):
ACK d213a64c8622a67d90da6a485d2b40f392edcecd

looks good in preview! I also tried resizing the browser window and everything resided as expected. If no objections I'll merge this tomorrow.

rajarshimaitra (approved):
ReACK d213a64c8622a67d90da6a485d2b40f392edcecd

PR #126, opened 2022-10-27
Add API docs for bdk-jvm and bdk-android 0.10.0

thunderbiscuit:
This PR updates both API reference websites available in the sidebar to their latest version, `0.10.0`.

It's a lot of files but really it's just replacing the directories in ./docs/.vuepress/public/android and ./docs/.vuepress/public/jvm.


notmandatory (approved):
ACK f598ea203d8819a6d163f36367ca082b64fffb17

PR #127, opened 2022-10-28
Fix path for Android and JVM API docs

thunderbiscuit:
Sorry I forgot to change the path for the API docs after changing their location in the `public/` directory.


notmandatory (approved):
ACK d35ef506dfe358e850cc94f52c285c652c8028e3

PR #133, opened 2022-12-07
Add API docs for bdk-jvm and bdk-android 0.25.0

thunderbiscuit:
This PR updates all three API reference websites available in the sidebar to their latest version, `0.25.0`.

It's a lot of files but really it's just replacing the directories in 
- `./docs/.vuepress/public/android`
- `./docs/.vuepress/public/jvm`
- `./docs/.vuepress/public/java`


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
notmandatory: Can you remove this file and add it to the .gitignore ?

thunderbiscuit: Done!

PR #134, opened 2022-12-17
Using BDK with Tor

rorp:
Added Tor tutorial

See https://github.com/bitcoindevkit/bdk_wallet/issues/196

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -60,7 +60,8 @@ const tutorialSidebar = [
       '/tutorials/using_bdk_with_hardware_wallets',
       '/tutorials/exploring_bdk_flutter',
       '/tutorials/bdk_cli_basics',
-      '/tutorials/bdk-cli_basics_multisig_2of3'
+      '/tutorials/bdk-cli_basics_multisig_2of3',
+      '/tutorials/bdk_with_tor',
     ],
   }
 ]
--- a/docs/tutorials/bdk_with_tor.md
+++ b/docs/tutorials/bdk_with_tor.md
@@ -1,0 +1,406 @@
+---
+title: "Using BDK with Tor"
+description: "How to integrate Tor client to sync BDK wallet with tor enabled blockchain service"
+authors:
+    - rorp
+date: "2023-01-03"
+tags: ["tutorial", "tor", "wallet", "blockchain"]
+---
+
+## Introduction
+
+It’s easy to underestimate the importance of privacy tech for Bitcoin, 
+especially when connecting to third party services. They can learn your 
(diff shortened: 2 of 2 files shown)


notmandatory:
Thanks for this, a few suggestions:

1. In the first section after explaining how to install tor you should probably also add that it needs to be started in another shell with:
   ```shell
    % tor
    ```
   And then need to wait for the boot strapping to finish.
2. Some of your code examples need the code indents fixed. 
3. You need to sign your commit.
4. You may want to remove the compact block filters example since it takes so long to sync, or at least start with something quicker like electrum, and then show it with cbf. 
5. Since this is a tutorial you should probably move it to the tutorial part of the website

notmandatory:
I've confirmed the CBF, electrum and both esplora sync parts work. But I did have to run the electrum and one of the esplora examples twice after seeing timeouts, most likely tor network related. So probably should mention that if the reader sees a timeout they should try rerunning it. 

I was unable to get the libtor example to build, I'm on an M1 mac and that crate may need to be updated to work with an aarch64 system.

rorp:
> I was unable to get the libtor example to build, I'm on an M1 mac and that crate may need to be updated to work with an aarch64 system.

`libtor` builds Tor from the sources, it's 100% compatible with aarch64 systems. Most likely you don't have autotools and/or libtool installed. I need to update the article with the build instructions.

notmandatory:
> > I was unable to get the libtor example to build, I'm on an M1 mac and that crate may need to be updated to work with an aarch64 system.
> 
> `libtor` builds Tor from the sources, it's 100% compatible with aarch64 systems. Most likely you don't have autotools and/or libtool installed. I need to update the article with the build instructions.

I fixed my build problem, I had to set these env variables, now builds fine:
```shell
export LDFLAGS="-L/opt/homebrew/opt/openssl@3/lib"
export CPPFLAGS="-I/opt/homebrew/opt/openssl@3/include"
```

notmandatory:
You'll also need to update the `config.js` file to add your tutorial to the sidebar list.

notmandatory (approved):
ACK 35a6e6432afae6249efbc6afea22e3c69414865e

All example code works for me, thanks for the updates. We will need to hold off on merging until https://github.com/bitcoindevkit/bdk/pull/816 goes in so the link to the example code is available.

thunderbiscuit:
Is this ready to go in? @rorp @notmandatory let's get it in if we can and the content is still relevant.

notmandatory:
@thunderbiscuit sorry for the slow response, yes let's merge this. It's great info and will be something we want to incorporate into a future user bdk user guide.

PR #135, opened 2023-01-09
docs: Bitcoin Core now supports miniscript inside wsh()

darosior:
As of 24.0(.1) Bitcoin Core now supports using Miniscript fragments inside `wsh()` (the only context for which Miniscript is fully defined).


notmandatory:
Concept ACK, I want to double check the data and then this can go in.

notmandatory (approved):
ACK 22bfc676bd7712db89925cf9526ff6cc3db9834a

PR #138, opened 2023-01-31
Update the playground to bdk-cli 0.27.0

danielabrozzoni:


playground/README.md
```diff
@@ -15,7 +15,7 @@ Compile bdk-cli using the right features and the WASM target:
 
 ```
 cd bdk-cli
-RUSTFLAGS="-C opt-level=s -C strip=symbols" cargo build --release --no-default-features --features esplora-reqwest,async-interface,compiler --target wasm32-unknown-unknown
+RUSTFLAGS="-C opt-level=s -C strip=symbols" cargo build --release --target wasm32-unknown-unknown --no-default-features --features esplora-reqwest,async-interface,compiler,dev-getrandom-wasm
 ```
```
notmandatory, line 20: Can we mention something like this too in the README too?
```suggestion

If building on Mac M1, due to bitcoin/rust-secp256k1#283 you must add to the build command `AR=/opt/homebrew/opt/llvm/bin/llvm-ar CC=/opt/homebrew/opt/llvm/bin/clang`.
```

notmandatory (approved):
tACK 8dd68cefd47b3ccb3a321071541e687a2326afbd

I tested the build process locally from the playground/README and made one comment, but otherwise all works great!

playground/README.md
danielabrozzoni: Updated, thanks!

notmandatory (approved):
re-tACK a1c27c32876ccb1780d5f528018556012c236b1d

PR #142, opened 2023-05-17
Update dependencies and build

dennisreimann:
As talked about with @ConorOkus, this upgrades to the new dev kit theme and updates the build.


notmandatory (approved):
LGTM. I verified playground, diagrams, links all work and generally looking around I didn't see any problems.

PR #153, opened 2023-06-30
link to docs.rs latest

garyray-k:
Update link to docs for bdk-cli so it navigates to latest and to a static verison


thunderbiscuit (changes requested):
This new link doesn't actually work well, sending me to a sort of in-between, bdk-cli version 0.27.1 (which doesn't exist).

I would merge an update to the latest cli version, [0.5.0](https://docs.rs/bdk-cli/0.5.0/bdk_cli/index.html).

PR #161, opened 2023-10-05
Add Bull Bitcoin to case studies page

thunderbiscuit:
This PR adds Bull Bitcoin and BitMask to the case studies page.

cc @BullishNode
cc @cryptoquick


danielabrozzoni:
This looks really good! A couple of nits:
- At the moment bitmask is out of center, I think we should use a 1:1 logo (similar to the mutiny one, with a transparent border on top and bottom)

Selection_230

- Doesn't have to be fixed here, but maybe the icons of the projects should have a max size? From desktop they look quite big, compared to, for example, the ldk website:
Selection_229
Selection_228

ConorOkus:
@thunderbiscuit great job on getting this in. I'd just say try to optimize the images to be something like 130*130. That's what we have over on the LDK side anyway

thunderbiscuit:
Thank you for the comments!

This PR now fixes the size of the images, the length of some of the taglines, and adds the 2 new projects.

notmandatory (approved):
Looks good! ACK b56cf1582982812018898ab3a85564d6d2edc6a4

PR #166, opened 2023-12-26
Correcting hello-world with update to faucet_address script_pubkey

upjohnc:
While working through the tutorial, there was a small error in the transaction builder.  It looks like the payload attribute of the Address has the script_pubkey function.  I figured, I'd submit the small change to help others in the future.

- the function `script_pubkey` is on the payload attribute of the Address

### DIFF
--- a/docs/tutorials/hello-world.md
+++ b/docs/tutorials/hello-world.md
@@ -130,7 +130,7 @@ let faucet_address = Address::from_str("mkHS9ne12qx9pS9VojpwU5xtRd4T7X7ZUt")?;
 
 let mut tx_builder = wallet.build_tx();
 tx_builder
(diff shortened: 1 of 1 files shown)


thunderbiscuit:
This tutorial is really old at this point, and I don't know that we have in mind to actually keep those as active pages or maybe have them in some sort of archive or something else. In any case, the way I interpret it, the post does not (although it should) define a specific version for which it was written, and simply points to the latest stable release of the library (0.29.0). This library uses rust-bitcoin 0.30.1, for which the `script_pubkey()` method is indeed defined on the `Address` type.

I'm closing this without merging because I think a user attempting this with the 0.29.0 version of the library should not run into the particular bug you are describing (which comes from earlier versions of rust-bitcoin), but let me know if I got something wrong, and feel free to reopen if need be!

PR #167, opened 2024-01-05
Add new projects to case studies page

thunderbiscuit:
This PR adds 4 new projects to the Case Studies page:
1. Stackmate
2. Lipa
3. Lexe
4. 10101


notmandatory (approved):
ACK 901ea2c87092458b25285875cbdbb25dc8e87ba8

I think 10101 could also fit the "exchange" category, but this is also fine as it is. I verified links works and text looks good.

PR #175, opened 2024-04-23
Resolve Windows build breaks

rustaceanrob:
While it is possible to configure SQLite for Windows, I spent a considerable amount of time on StackExchange researching linker errors and Windows configs. Even though I don't often use Windows I can imagine this being a deterrent for people just trying to get familiar with BDK. The added CLI command built successfully on Windows for me.


thunderbiscuit (approved):
Awesome. Thank you on behalf of Windows users, who are so often forgotten in this field. I don't have a way to test this myself, but I'll take your word for it.

PR #177, opened 2024-05-02
Refactor of case studies page

thunderbiscuit:
The case studies page is really cool, and it's getting a little busy. This PR is looking at revamping it a little bit. Here are a few things that I thought could use some polishing. Most of those are orthogonal to each other, so they could be applied separately, i.e. this is not a package PR, and we can cut an choose.
- The navbar name "Case Studies" was better suited for earlier stages, where people were looking for a few examples of applications of bdk in production. Those projects became "case studies". As time goes on a our list of in-prod applications is up to 23 and growing, these a not just case studies anymore. They're users of bdk, builders of products people use across the globe on variety of platforms.
- In order to keep the page looking clean, the projects could only really use 5-12 word sentences, otherwise the alignment gets all out of sync (take a look at the Envoy wallet on an 800px browser for example). It would be great to allow slightly more verbose explanations of the project. The new, vertical list allows that better.
- The use of the sidebar for the project grouping is clean and simpler on the eye, particularly because it follows the layout of other navbar links.

You can compare
- [The current page](https://bitcoindevkit.org/case-studies/)
- [The new pages proposed in this PR](https://deploy-preview-177--awesome-golick-685c88.netlify.app/buildingwithbdk/all/)

TODO
- [x] Introduce new rules for responsiveness on smaller screens
- [x] All pages for the different groupings

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -45,6 +45,23 @@ const docsSidebar = [
   }
 ]
 
+const builtWithBdkSidebar = [
+  {
+    title: 'Built With BDK',
+    collapsable: false,
+    children: [
+      ["/adoption/all.md", "All"],
+      ["/adoption/mobile.md", "Mobile"],
+      ["/adoption/desktop.md", "Desktop"],
+      ["/adoption/hardware.md", "Hardware"],
+      ["/adoption/web.md", "Web"],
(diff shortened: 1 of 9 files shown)


notmandatory (approved):
ACK https://github.com/bitcoindevkit/bitcoindevkit.org/commit/7d2365bdb3ee4ab642716ecc6169944684075d68

PR #179, opened 2024-05-09
fix: supporters link on main page

reez:
I was getting a 404 when tapping the Supporters link under "More", I think this fixes the path for it.

Screenshot 2024-05-09 at 4 00 13 PM


thunderbiscuit (approved):
ACK.

PR #184, opened 2024-05-21
blog: add 2024 Q1 project update

notmandatory:


thunderbiscuit (approved):
ACK f0a73cf169dd864f86c2ff2ab0da17fb1bf23274.

I'm not sure why the title is so close to the metadata in practice, but if you wanted you could add a <br> before the first header and hope that creates a bit of spacing. Not a blocker, it's fine as is.

PR #185, opened 2024-06-20
Create devcontainer.json

Shiranm88:


notmandatory:
There's no explanation why this change is needed so I'm closing it. Feel free to update it with more details if you still think this change is needed.

PR #187, opened 2024-07-05
.github/workflows/build.yml:build matrix

RandyMcMillan:


.github/workflows/build.yml
```diff
@@ -3,6 +3,9 @@ name: Build
 on:
   # Run the build for pushes and pull requests targeting master
   push:
+    tags:
```
thunderbiscuit, line 6: Can you tell me what the purpose of this is?

.github/workflows/build.yml
```diff
@@ -14,14 +17,17 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+    strategy:
```
thunderbiscuit, line 20: You're using a matrix but then only setting one version, so it's adding complexity that's not used. Why not simply keep the `node-version: 16` as before?

thunderbiscuit (commented):
The changes in `installation.md` seem unrelated to this PR (but then again we don't have an explanation for what this PR attempts to do so maybe it was the core of it?)

Let us know what you mean to do when opening PRs, just so we can review against that.

thunderbiscuit:
Overall my question here is: what is this PR doing that is bettering the codebase?

thunderbiscuit:
This PR is stale and doesn't have a clear objective so I'll close it for now, but Randy feel free to reopen if you want to explain a bit more what it was about and why we should merge!

PR #189, opened 2024-08-09
Add 2024 Q2 update blog post

thunderbiscuit:
I used to have a workflow to build the site locally but it doesn't work anymore so I don't have a way to test this. Pushing here so I can see the preview.

Edit: ok it's not just on my machine.

So the key is to add
```js
function capitalize(str) {
  return str.charAt(0).toUpperCase() + str.slice(1);
}
```

To the `node_modules/@spiralbtc/vuepress-devkit-theme/config.js` file to enable it to build locally. For deployment, we'd need to update the theme itself, but I can't find the github link! @dennisreimann any ideas?

PR #190, opened 2024-08-09
Upgrade dependencies

dennisreimann:
Fixes the build issues in #189.

@thunderbiscuit Can you check if this resolves the build error for you and if so merge this and then rebase your blog PR on top of the fix?

@ConorOkus updated the spiral theme and the latest version already includes the missing function. For reference: The theme is maintained here: https://github.com/spiralxyz/vuepress-devkit-theme

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,13 +9,13 @@
       "version": "1.0.0",
       "license": "MIT",
       "devDependencies": {
-        "@spiralbtc/vuepress-devkit-theme": "^0.5.0",
+        "@spiralbtc/vuepress-devkit-theme": "^0.19.0",
         "broken-link-checker": "0.7.8",
-        "serve": "14.2.0",
-        "start-server-and-test": "2.0.0",
-        "vue": "2.7.14",
-        "vue-server-renderer": "2.7.14",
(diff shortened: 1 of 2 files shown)


thunderbiscuit:
Wow thanks for the super quick fix!

ACK a5317cc58a6b794bc49fc2ea2f4e03502ff4cb65. I see why I couldn't find it; I'm not part of the spiralxyz org. All is good now!

thunderbiscuit (approved):

thunderbiscuit:
Ok actually it only worked locally because I had added the method to my local theme. When I remove the function it still doesn't build, even though my package-lock shows 
```json
"@spiralbtc/vuepress-devkit-theme": "^0.19.0",
```

dennisreimann:
Did you run `npm install` to upgrade the dependencies locally?

thunderbiscuit:
I had not! Thanks. But actually even after doing that I get the same error. Both my package.json and package-lock.json use the `"@spiralbtc/vuepress-devkit-theme": "^0.19.0",` dependency.

It also appears that it's breaking for Netlify (#189).

Issue #98, opened 2022-04-25
Make local docs links relative


notmandatory:
Would be good to figure out how to make local rust nightly, android and kotlin jvm docs links relative. As they are now they point to existing public URLs so if docs are changed in the PR then the preview won't point to the new pages.

_Originally posted by @notmandatory in https://github.com/bitcoindevkit/bitcoindevkit.org/pull/96#discussion_r857878062_

thunderbiscuit:
This is fixed by #226.

PR #226, opened 2025-03-26
Update documentation page

thunderbiscuit:
Small fixes to the documentation page, including a fix for #98.

### DIFF
--- a/docs/docs/README.md
+++ b/docs/docs/README.md
@@ -1,12 +1,12 @@
 # Documentation
 
-## Book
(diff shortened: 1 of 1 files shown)


thunderbiscuit:
Note that those API docs websites are internal to the website, so the standard markdown links do not open new tabs for them. We have to use direct html for the language bindings API docs.

reez (approved):
ACK 06fa6dc9ef227ce46e21394db683daf79118405b

This is a nice update

PR #203, opened 2024-11-08
(draft) fix: q3 update date

reez:
https://bitcoindevkit.org/blog/_2024-q3-update/ shows a date of 10/32/2024 

Screenshot 2024-11-08 at 10 38 58 AM
Screenshot 2024-11-08 at 10 39 03 AM

So trying to make a fix for that

PR #230, opened 2025-04-29
chore: fix date on 2025_q1_update

reez:
Screenshot 2025-04-28 at 4 16 30 PM

Date on blog shows ^

Date with this pr fix shows this now v

Screenshot 2025-04-28 at 8 10 12 PM

I remember running into it once before and doing a fix https://github.com/bitcoindevkit/bitcoindevkit.org/pull/203, no idea why this is happening maybe some weird bug in the template (somehow bugs out on dates that are the first of the month sometimes?), will follow back up on maybe fixing the template or tracking that down another time.

For now this fixes, tested locally and also the deploy preview looks good to me.


thunderbiscuit (approved):
ACK.

PR #240, opened 2025-07-17
foundation: about 501(c)(6)

reez:
we finally have our official 501(c)(6)  "positive determination" letter from the IRS. we can now update the wording on our foundation page to explicitly call out our status.


thunderbiscuit (approved):
ACK 513ff50fddf07aba697da52452df9431b3d76bc2.

PR #242, opened 2025-07-22
docs: add youtube link

reez:
On the recent BDKF call it was mentioned to add BDK YouTube link to the website, which was a great idea.

https://bitcoindevkit.org/docs/ page seemed like the best spot. Originally it was structured as 3 sections:
- The Book of BDK
- Rust API Documentation
- Language Bindings API Documentation

I was trying to figure out the best spot to put BDK YouTube in that structure. The more I thought about it I landed on an idea of changing the first section to "Guides" which would include Book of BDK + BDK YouTube (since they seemed like they logically could be grouped together nicely), so that's what this PR currently includes. Structure would be 3 sections:
- Guides
- Rust API Documentation
- Language Bindings API Documentation

The other option would be to create a separate "Videos" section with just BDK YouTube in it, but I just didn't like that idea as much at the moment. That would have the structure as:
- The Book of BDK
- Videos
- Rust API Documentation
- Language Bindings API Documentation

But I don't really have a strong opinion so looking for feedback from ya @notmandatory @thunderbiscuit because I'm cool with whatever I'm not dead set on either of the options I thought of.

### DIFF
--- a/docs/docs/README.md
+++ b/docs/docs/README.md
@@ -1,10 +1,11 @@
 # Documentation
 
-## The Book of BDK
+## Guides
 
-The "Book of BDK" is a gentle introduction to using the BDK suite of libraries. It includes a _Getting Started_ guide and _Cookbook_ section with example code in Rust, Kotlin and Swift. It is still a work in progress, contributions welcome.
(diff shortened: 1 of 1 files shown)


thunderbiscuit (approved):
This is good!

ACK da469a2c5c296021d5d0ab55bd65b697c4764d92.

reez:
Thanks thunder!!!

notmandatory:
post merge ACK, looks great!

PR #247, opened 2025-10-15
chore: move grantee from current to past

reez:
Move grantee from current to past


thunderbiscuit (approved):
ACK 355315f6e2dfe8ca7adfa94f53ab9dbb769febc9.

PR #251, opened 2025-12-01
Update Bitcoin-Safe details

design-rrr:
cc: @andreasgriffin


reez:
Will let Andreas weight in on this change.

Also here is the deploy preview which probably isn't what we want before merge:

Screenshot 2025-12-01 at 3 28 04 PM

design-rrr:
Thanks for sharing the preview. Added some attributes to the images that should fix the dimension.

andreasgriffin:
> Will let Andreas weight in on this change.

@design-rrr  Thanks for updating the logo and slogan.   Looks good!

reez (approved):
ACK 5c71a1b363b05502bd3e0995b9334f14c3c1bf8b

PR #276, opened 2026-04-24
Fix grantee social media icons, sort pgp keys and grantees

notmandatory:
* Fix grantee social media icons to use `material-xxx` the old `lucide-` ones don't work any more
* Sort PGP keys alphabetically, put security key on top, removed Riccardo, Joe and Wei; add Tobin
* Sort grantees page alphabetically


reez (approved):
ACK 3dc9939a03ae2081db762048dc745e4508545caf

PR #280, opened 2026-04-29
Update Rob's description

rustaceanrob:
If someone has read this deeply into the site, they should be aware that they can experiment and test compact block filters via BDK supported apps.


thunderbiscuit (approved):
ACK f1dd63cd7474726e48162cc1b54a07395279f52f. Thanks!

PR #286, opened 2026-06-12
fix: correct grantee section structure

reez:
Fixes the grantees page section structure so ValuedMammal’s links and bio appear under the `## ValuedMammal` heading instead of being included in the `## thunderbiscuit` section.


notmandatory (approved):
ACK thanks!

